PR #3, opened 2019-09-21
add n64 driver

vieux:


sezero:
Thank you!

1. Nitpicking: N64_Init() should return int, not BOOL. (can fix
   it after myself I merge.)

2. About the workaround added to mmio.c for LONG_MAX:  I am not
   comfortable with it. Does your SDK not have a limits.h (your
   CFLAGS in your makefile doesn't have -DHAVE_LIMITS_H)? Or does
   it not define LONG_MAX (how??)

vieux:
Sure, let me fix 1. as soon as I'm back on a computer. Regarding 2. I'll investigate.

sezero:
FWIW, I just made simplifications: see attached patch.
It builds for me using the old precompiled windows-hosted
toolchain from 2009 at https://dragonminded.com/n64dev/

[simplified_patch.txt](https://github.com/sezero/mikmod/files/3639448/simplified_patch.txt)

vieux:
I pushed a few changed based on your patch, thanks. one question though.

I used the libdragon's  examples as template for the makefile, I'm not sure having `$(N64_INST)/bin` in your PATH is common practice with libdragon, could we do with those changes ?

sezero:
OK, merged it as it is.  Thanks!

vieux:
@sezero oh thanks for the quick merge, just to be sure, the current state of the commit is that you do `cd n64 &&  make -f Makefile.n64` is that consistent with the other Makefiles ?

sezero:
> `cd n64 && make -f Makefile.n64` is that consistent with the other Makefiles ?

Yes.

sezero:
Just added -DHAVE_SNPRINTF to the makefile, you may want that.
(commit 600b62f9e993570c7de26d642eac36e3a91bf139)

sezero:
Pushed two or three further commits - you may want to pull.

sezero:
Pulled from your branch and pushed a minor simplification to N64_Update().
Please make sure things are OK, so that I can consider this as stable.

vieux:
Cool thanks, I was about to send a PR for the can write. I'll make one last
test and report back here

vieux:
@sezero unrelated to the last changes, but I do see a warning during compilation:

```
../loaders/load_669.c: In function 'S69_Load':
../loaders/load_669.c:297:4: warning: 'strncpy' output may be truncated copying 36 bytes from a string of length 107 [-Wstringop-truncation]
  297 |    strncpy(of.comment,mh->message,36);
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

do you see the same on other platforms ?

sezero:
> do you see the same on other platforms ?

Yes, and ignoring it.

vieux:
ok 👍 I was able to compile my game with the latest commit and it seems to work great

sezero:
Great!   I consider the n64 addition as stable then.  Thanks.

PR #4, opened 2020-05-14
Fix attribution on N64 driver

DragonMinded:
Stumbled on this, I wrote the N64 code way back in 2009 and I guess it got uploaded to mediafire and put on n64dev.net?


sezero:
Thanks.

PR #18, opened 2020-10-26
Fix MED 9xx command in tempo mode and various BPM mode bugs.

### LINKED ISSUE #17: .MED secondary tempo (9xx) command is bugged.
Creating an issue for this one so my research doesn't clutter the (eventual) PR.

This bug occurs with .MED files that use the .MOD secondary tempo command `9xx` in tempo mode, and appears to be caused by this line of code: https://github.com/sezero/mikmod/blob/master/libmikmod/loaders/load_med.c#L223
```C
	  case 0x9:
		if (bpmtempos) {
			if (!dat)
				dat = of.initspeed;
			UniEffect(UNI_S3MEFFECTA, dat);
		} else {
			if (dat <= 0x20) {
				if (!dat)
					dat = of.initspeed;
				else
					dat /= 4; /* This line. */
				UniPTEffect(0xf, dat);
			} else
				UniEffect(UNI_MEDSPEED, ((UWORD)dat * 125) / (33 * 4));
		}
		break;
```
The OctaMED documentation suggests that this should convert directly to/from the .MOD speed command and, considering MikMod's tempo conversion, it seems that there is no reason to divide this value by 4. https://modland.com/pub/documents/format_documentation/Music%20Editor%20v3.2%20(.med).txt

This change dates back to MikMod 3.1, where it appeared in the form of a `speeddivisor` variable that was always equal to 4 during tempo mode. This variable is applied to the `9xx` command and the speed case of the `Fxx` command in different ways!
```
- Lots of bug fixes in MED loader. Should now play modules at correct speed, but
  still not perfect
```
```C
                case 0x9:
                        if(dat<=0x20) {
                                if(!dat) currentspeed=of.initspeed;
                                else {
                                        if(bpmtempos) currentspeed=dat*4/speeddivisor;
                                        else currentspeed=dat/speeddivisor;
                                }
                                UniPTEffect(0xf,currentspeed);
                        } else ...
```
```C
                case 0xf:
                        switch(dat) {
...
                                default:
                                        if(dat<=10) {
                                                currentspeed=dat*4/speeddivisor;
                                                UniPTEffect(0xf,currentspeed);
                                        } else
...
```
By 3.1.5 this variable was completely removed, just leaving a fixed division by 4 in the `9xx` command and not in the `Fxx` command. This looks like an oversight that never got caught to me, but there might some reason I'm missing for this being there.

Removing that line makes these example .MEDs play at the same speeds that xmp plays them at. I don't have OctaMED set up to test with right now.

[alain.med.zip](https://github.com/sezero/mikmod/files/5415282/alain.med.zip)
[wizardry.med.zip](https://github.com/sezero/mikmod/files/5415283/wizardry.med.zip)
[yellybeansong.med.zip](https://github.com/sezero/mikmod/files/5415284/yellybeansong.med.zip)

Since this division by 4 was presumably added for a reason (even if it might have been a mistake), I'm going to exhaustively test (with OctaMED, xmp, and MikMod) every .MED MMD0 and MMD1 file I can find that uses this to make sure removing the bugged line doesn't cause any breakage.

edit: maybe that was too optimistic, there are 379 MMD0 files and 816 MMD1 files that use `9xx` with values <= `0x20`...

AliceLR:
* 9xx with values <=0x20 in tempo mode now sets the correct speed.
* 9xx with values >0x20 in either mode now is ignored.
* Fxx with values from 01 to 0A in tempo mode now sets the primary
  tempo like it's supposed to (via a lookup table). A similar bug in
  the initial tempo was also fixed.
* Fxx with values from 01 to 0A in BPM mode now set the BPM.
* Fxx in BPM mode now respects the rows per beat value.
* OctaMED has a weird bug when setting the BPM to 1 or 2 that is used
  at the end of electrosound 64.med. A check for this has been added.
* The initial speed and initial tempo are no longer divided by
  GCD(initial speed, rows per beat) in BPM mode, fixing numerous bugs.
* Added support for legacy 8-channel mode tempos.
* Fixed a bug where patterns with fewer tracks than the total number
  of tracks wouldn't be loaded correctly.

Fixes #17.

### DIFF
--- a/libmikmod/loaders/load_med.c
+++ b/libmikmod/loaders/load_med.c
@@ -151,6 +151,8 @@ static MMD1NOTE *mmd1pat = NULL;
 
 static BOOL decimalvolumes;
 static BOOL bpmtempos;
+static BOOL is8channel;
+static UWORD rowsperbeat;
 
 #define d0note(row,col) mmd0pat[((row)*(UWORD)of.numchn)+(col)]
 #define d1note(row,col) mmd1pat[((row)*(UWORD)of.numchn)+(col)]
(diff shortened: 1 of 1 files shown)


sezero:
Thanks.

Issue #21, opened 2020-10-29
MOD retrigger effect is broken.


AliceLR:
The .MOD (and .XM, .MED, and whatever else uses it...) retrigger effect currently does not initialize the retrigger timer, meaning it will always retrigger on the second tick of a given row.

Test file: in Protracker 3.15, libxmp, OpenMPT, this plays one snare hit, then three evenly spaced snare hits. In MikMod, this plays one snare hit, then four unevenly spaced snare hits. Speed=24, uses `E98`.

[retrig.mod.zip](https://github.com/sezero/mikmod/files/5455898/retrig.mod.zip)

The only relevant changelog entry I could find re: this was for libmikmod-3.1.10:
```
- ProTracker effect E9 (Retrig) was not played correctly.
```

But weirdly, the only change to this effect between libmikmod-3.1.9 and libmikmod-3.1.10 is to prevent it from triggering on tick 0:
libmikmod-3.1.9:
```C
                case 0x9: /* retrig note */
                        /* only retrigger if data nibble > 0 */
                        if (nib) {
                                if (!a->retrig) {
                                        /* when retrig counter reaches 0, reset counter and restart
                                           the sample */
                                        if (a->period) a->kick=KICK_NOTE;
                                        a->retrig=nib;
                                }
                                a->retrig--; /* countdown */
                        }
                        break;
```
libmikmod-3.1.10:
```C
        case 0x9: /* retrig note */
                /* do not retrigger on tick 0, until we are emulating FT2 and effect
                   data is zero */
                if (!tick && !((flags & UF_FT2QUIRKS) && (!nib)))
                        break;
                /* only retrigger if data nibble > 0, or if tick 0 (FT2 compat) */
                if (nib || !tick) {
                        if (!a->retrig) {
                                /* when retrig counter reaches 0, reset counter and restart
                                   the sample */
                                if (a->main.period) a->main.kick=KICK_NOTE;
                                a->retrig=nib;
                        }
                        a->retrig--; /* countdown */
                }
                break;
```
The worst part of this is that I loaded a test .MOD into Protracker 3.15 and it *does* retrigger on tick 0.

Test file: this plays one snare, then plays a single snare note with an `E9A` effect on the note line and an `E9A` effect on the following line. In Protracker, the two lines with `E9A` both play on ticks 0, 10, and 20. libxmp, MikMod, OpenMPT, and MPT 1.16 all do something else(!).

[...]

PR #22, opened 2020-10-29
Fix E9x effect bug that would cause it to always retrigger on tick 1.

AliceLR:
This patch fixes the overtly wrong behavior of the .MOD `E9x` retrigger effect and attempts to at least establish a sensible default behavior in the spirit of what the original code here was supposed to do. As mentioned by @sagamusix in #21 this probably will need to be followed up with careful inspection/correction for individual formats and .MOD variants (I commented a list of different affected formats also in that issue).


sezero:
Merged.  I hope that this doesn't open a can of worms :)

AliceLR:
I can't imagine this made anything worse off than it was before... :(

sezero:
OK :)

PR #25, opened 2020-11-14
Add support for MED song/instrument transpose and octave wrapping.

### LINKED ISSUE #24: OctaMED transpose and octave wrapping are not supported.
Several behaviors of OctaMED affecting note translation do not currently work:

* Song transpose: when this is set, every note in a song is transposed a number of semitones up or down. Used by around 80 MMD0s/MMD1s.
* Instrument transpose: when this is set, every note played by a particular instrument is transposed a number of semitones up or down. Used by around 750 MMD0s/MMD1s, more than any other unimplemented OctaMED feature so far (aside from MMD2/3...).
* Octave wrapping: hardware mixing mode has a couple of strange behaviors that a few modules rely on: octaves 4-7 act like octave 3; octaves 8-A wrap around to two octaves below octave 1 due to a bug. I don't have numbers on the former behavior but the latter affects ~30 MMD0s/MMD1s maximum, likely fewer.

The fix for all of these is the same: the OctaMED loader should allocate instruments and implement these features with the tuning table. This is a fairly easy fix (and this will be required to implement the IFFOCT instruments down the road anyway).

Tentative patch here. It is not well-tested yet, so expect some table spam here as I do that over the next few days.
https://github.com/sezero/mikmod/compare/master...AliceLR:med-transpose-octave-wrap

AliceLR:
Implements both OctaMED transpose types, and also OctaMED's glitchy octave range behavior, by allocating instruments for MED modules. This should also be helpful in the future for implementing IFFOCT instruments.

Fixes #24.

### DIFF
--- a/libmikmod/loaders/load_med.c
+++ b/libmikmod/loaders/load_med.c
@@ -48,6 +48,8 @@ extern int fprintf(FILE *, const char *, ...);
 
 /*========== Module information */
 
+#define MEDNOTECNT 120
+
 typedef struct MEDHEADER {
 	ULONG id;
 	ULONG modlen;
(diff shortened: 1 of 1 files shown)


sezero:
Thanks.

PR #26, opened 2020-11-15
Use more accurate BPM values for 8-channel mode tempos.

AliceLR:
As noted in #24, there are some accuracy issues with the 8-channel mode tempo table copied from libxmp (which appears to be derived from OctaMED SS 1.03c's tempo conversions). This adds a new table of BPMs for 8-channel mode that are much closer to OctaMED playback (see https://github.com/libxmp/libxmp/issues/232).

Marking this as a draft since the linked libxmp issue and measurement of these BPMs should probably at least be checked first, but since MikMod doesn't have non-integer BPMs or anything like libxmp's time factor this table is probably as close as it can get.

### DIFF
--- a/libmikmod/loaders/load_med.c
+++ b/libmikmod/loaders/load_med.c
@@ -227,26 +227,29 @@ static UWORD MED_ConvertTempo(UWORD tempo)
 	};
 
 	/* MEDs with 8 channels do something completely different with their tempos.
-	   This behavior completely overrides BPM mode timing when it is enabled.
-	   Table stolen from libxmp ;-( */
+	   This behavior completely overrides normal timing when it is enabled.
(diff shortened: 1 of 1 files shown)


sezero:
What's the status of this patch?

AliceLR:
I'm not sure how correct this patch is. I think these BPM tables are about as reasonably close to the original speeds as can be done with integer BPMs, but 1) I was hoping someone else familiar with OctaMED could check my work, and 2) there's the minor problem that newer OctaMED Soundstudio versions have a different BPM table. The SS 2 table may simply be the rounded averages of the two tables I measured, which would work around the issue that there are two different tables and still be reasonably close enough to the correct BPMs for the affected modules to sound correct.

AliceLR:
I went ahead and just used the OctaMED Soundstudio 2 timing for this.

sezero:
Thanks.

PR #27, opened 2020-11-17
Fix GDM set pan position effect.

AliceLR:
...yeah, this was broken here too, due to being implemented in a confusing way that never worked.

See also: https://github.com/libxmp/libxmp/pull/236

edit: I think set global volume is also bugged but that can be a separate patch once I verify that.


sezero:
Thanks.

PR #28, opened 2020-11-18
Merge vibrato tables/math, fix tremolo ramp down, fix IT vibrato.

AliceLR:
This is a draft for now because I need to do a little more research to finish it up, but as-is I think any behaviors explicitly changed should be correct and everything else should be equivalent to before.

* `mplayer.c` had two separate vibrato tables for effect vibrato and autovibrato. Effect vibrato used a table of size 32 and effective period of 64,  but autovibrato had a table of size 128 (but with 1/4 the amplitude precision) and effective period of 256. I replaced them with a single table of size 128 and amplitudes from 0 to 255. I can't imagine there was any reason for making effect vibrato have lower precision than autovibrato.
* All of the commands using `VibratoTable` had the same waveform switch and math copypasted inline. Since most compilers newer than 1998 can automatically inline small functions I combined these to clean up some code.
* The one place these were different was the IT implementations of vibrato (Hxy) and fine vibrato (Uxy), which had the "ramp down" and "square" waveforms flipped. I verified with Impulse Tracker 2.14 that these should always have been the same IDs as their MOD/S3M equivalents, so I'm not sure why this was changed in the first place (might need a little research to determine why they were flipped...).
* Every waveform handler was computing "ramp down" as a ramp up, which is correct for vibrato (period ramp up results in pitch ramp down) but caused tremolo to ramp the wrong direction. There is a separate waveform function for tremolo to address this.

To do:
- [x] ~~I did not change anything in particular about autovibrato aside from the table it uses but it is definitely much weaker than it ought to be for Impulse Tracker modules, has square at index 1 (need to check whether this is correct), and has two different ramps in indices 2 and 3 (IT should have a random waveform). This probably needs to be handled carefully since this behavior might be derived from XM.~~
- [x] Fix XM sine autovibrato.
- [x] Fix IT autovibrato waveforms.
- [x] Fix IT autovibrato depth.
- [x] Fix random waveform.
- [x] ~~The ramp flipping flag is kind of a hack, and doesn't account for the other waveforms that might be flipped entirely for vibrato as well. This needs some more verification. edit: for what it's worth, both XM and IT autovibrato flips every waveform.~~ Confirmed that all IT new vibrato effects subtract their waveforms from the period.
- [x] ~~MikMod appears to be checking the wrong bit to determine whether to use the continuous variant of a given waveform.~~ It was, fixed now.
- [x] MikMod was applying the MOD quirk of ignoring vibrato on the first tick to everything using `DoVibrato`. This is now a flag specified by whatever uses it, and S3M/XM/old effects IT/MED now ignore it.
- [x] Added commands for old effects IT. The IT vibrato effects now all use `DoITVibrato` to avoid adding even more copypaste.
- [x] Panbrello had the same reversed waveforms issue as IT vibrato and also wasn't correctly resetting the waveform position at the start of a note. Random panbrello does something different from everything else that requires a new variable in `MP_CONTROL` (not sure if that's an ABI change...) so it's just broken for now.

### DIFF
--- a/libmikmod/include/mikmod_internals.h
+++ b/libmikmod/include/mikmod_internals.h
@@ -341,6 +341,11 @@ enum {
  /* Last effect supported by old modules in the UNI format. */
     UNI_FORMAT_LAST,
 
+ /* Scream Tracker effects */
+    UNI_S3MEFFECTH,    /* vibrato */
+ /* Impulse Tracker effects */
+    UNI_ITEFFECTH_OLD, /* vibrato (old) */
+    UNI_ITEFFECTU_OLD, /* fine vibrato (old) */
  /* OctaMED effects. */
     UNI_MEDEFFECT_VIB, /* MED vibrato */
     UNI_MEDEFFECT_FD,  /* set pitch */
--- a/libmikmod/playercode/mlutil.c
+++ b/libmikmod/playercode/mlutil.c
@@ -182,9 +182,12 @@ void S3MIT_ProcessCmd(UBYTE cmd, UBYTE inf, unsigned int flags)
 					UniEffect(UNI_ITEFFECTG,inf);
 				break;
 			case 8: /* Hxy vibrato */
-				if (flags & S3MIT_OLDSTYLE)
-					UniPTEffect(0x4,inf);
-				else
+				if (flags & S3MIT_OLDSTYLE) {
+					if (flags & S3MIT_IT)
+						UniEffect(UNI_ITEFFECTH_OLD,inf);
+					else
+						UniEffect(UNI_S3MEFFECTH,inf);
+				} else
 					UniEffect(UNI_ITEFFECTH,inf);
 				break;
 			case 9: /* Ixy tremor, ontime x, offtime y */
@@ -197,9 +200,12 @@ void S3MIT_ProcessCmd(UBYTE cmd, UBYTE inf, unsigned int flags)
 				UniPTEffect(0x0,inf);
 				break;
 			case 0xb: /* Kxy Dual command H00 & Dxy */
-				if (flags & S3MIT_OLDSTYLE)
-					UniPTEffect(0x4,0);
-				else
+				if (flags & S3MIT_OLDSTYLE) {
+					if (flags & S3MIT_IT)
+						UniEffect(UNI_ITEFFECTH_OLD,0);
+					else
+						UniEffect(UNI_S3MEFFECTH,0);
+				} else
 					UniEffect(UNI_ITEFFECTH,0);
 				UniEffect(UNI_S3MEFFECTD,inf);
 				break;
@@ -259,9 +265,12 @@ void S3MIT_ProcessCmd(UBYTE cmd, UBYTE inf, unsigned int flags)
 				}
 				break;
 			case 0x15: /* Uxy Fine Vibrato speed x, depth y */
-				if(flags & S3MIT_OLDSTYLE)
-					UniEffect(UNI_S3MEFFECTU,inf);
-				else
+				if(flags & S3MIT_OLDSTYLE) {
+					if (flags & S3MIT_IT)
+						UniEffect(UNI_ITEFFECTU_OLD,inf);
+					else
+						UniEffect(UNI_S3MEFFECTU,inf);
+				} else
 					UniEffect(UNI_ITEFFECTU,inf);
 				break;
 			case 0x16: /* Vxx Set Global Volume */
--- a/libmikmod/playercode/mplayer.c
+++ b/libmikmod/playercode/mplayer.c
@@ -54,6 +54,11 @@ MODULE *pf = NULL;
 
 #define	HIGH_OCTAVE		2	/* number of above-range octaves */
 
+enum vibratoflags {
+	VIB_PT_BUGS	= 0x01, /* MOD vibrato is not applied on tick 0. */
+	VIB_TICK_0	= 0x02, /* Increment LFO position on tick 0. */
+};
+
 static	const UWORD oldperiods[OCTAVE*2]={
 	0x6b00, 0x6800, 0x6500, 0x6220, 0x5f50, 0x5c80,
 	0x5a00, 0x5740, 0x54d0, 0x5260, 0x5010, 0x4dc0,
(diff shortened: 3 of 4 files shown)


AliceLR:
> I'm not sure why this was changed in the first place (might need a little research to determine why they were flipped...).

This change occurred between mikmod 3.0.3 and mikmod 3.0.4 with no relevant changelog entry to document it.

3.0.3
```C
static void DoITVibrato(void)
{
    UBYTE q;
    UWORD temp;

    q = (a->vibpos>>2)&0x1f;

    switch(a->wavecontrol&3)
    {   case 0: /* sine */
           temp=VibratoTable[q];
        break;

        case 1: /* ramp down */
           q<<=3;
           if(a->vibpos<0) q=255-q;
           temp=q;
        break;

        case 2: /* square wave */
           temp=255;
        break;

        case 3: /* evil random */
           temp = rand() & 255;  /* (range 0 to 255) */
        break;
    }

    temp*=a->vibdepth;
    temp>>=8;
    temp<<=2;

    if(a->vibpos>=0)
       a->period = a->tmpperiod+temp;
    else
       a->period = a->tmpperiod-temp;

    a->vibpos+=a->vibspd;
}

static void DoITFineVibrato(void)
{
    UBYTE q;
    UWORD temp;

    q = (a->vibpos>>2)&0x1f;

    switch(a->wavecontrol&3)
    {   case 0: /* sine */
           temp=VibratoTable[q];
        break;

        case 1: /* ramp down */
           q<<=3;
           if(a->vibpos<0) q = 255-q;
           temp = q;
        break;

        case 2: /* square wave */
           temp = 255;
        break;

        case 3: /* evil random */
            temp = rand() & 255;  /* (range 0 to 255) */
        break;
    }

    temp*=a->vibdepth;
    temp>>=8;

    if(a->vibpos>=0)
       a->period = a->tmpperiod+temp;
    else
       a->period = a->tmpperiod-temp;

    a->vibpos+=a->vibspd;
}
```

3.0.4
```C
void DoITVibrato(void)
{
        UBYTE q;
        UWORD temp=0;

        q = (a->vibpos>>2)&0x1f;

        switch(a->wavecontrol&3) {
                case 0: /* sine */
                        temp=VibratoTable[q];
                        break;
                case 1: /* square wave */
                        temp=255;
                        break;
                case 2: /* ramp down */
                        q<<=3;
                        if(a->vibpos<0) q=255-q;
                        temp=q;
                        break;
                case 3: /* random */
                        temp = random() & 255;  /* (range 0 to 255) */
                        break;
        }

        temp*=a->vibdepth;
        temp>>=8;
        temp<<=2;

        if(a->vibpos>=0)
                a->period = a->tmpperiod+temp;
        else
        a->period = a->tmpperiod-temp;

        a->vibpos+=a->vibspd;
}

void DoITFineVibrato(void)
{
        UBYTE q;
        UWORD temp=0;

        q = (a->vibpos>>2)&0x1f;

        switch(a->wavecontrol&3) {
                case 0: /* sine */
                        temp=VibratoTable[q];
                        break;
                case 1: /* square wave */
                        temp = 255;
                        break;
                case 2: /* ramp down */
                        q<<=3;
                        if(a->vibpos<0) q = 255-q;
                        temp = q;
                        break;
                case 3: /* random */
                        temp = random() & 255;  /* (range 0 to 255) */
                        break;
        }

        temp*=a->vibdepth;
        temp>>=8;

        if(a->vibpos>=0)
                a->period = a->tmpperiod+temp;
        else
                a->period = a->tmpperiod-temp;

        a->vibpos+=a->vibspd;
}
```

AliceLR:
>  I did not change anything in particular about autovibrato aside from the table it uses but it is definitely much weaker than it ought to be for Impulse Tracker modules...

It uses some math I find questionable, but I suspect I'll have to further dismantle it (like I did with libxmp) to really figure out what's going on. XM autovibrato is not affected because it uses a different algorithm. An early changelog entry claimed that it was supposed to be correct for fine slides mode but it seems very weak in both modes.

> ...has square at index 1 (need to check whether this is correct), and has two different ramps in indices 2 and 3...might be derived from XM.

It is, these are the correct waveforms for XM autovibrato. IT should use the regular waveforms and XM should use those.

AliceLR:
Re: IT autovibrato from the MikMod 3.0.3 changelog:
```
 - ImpulseTracker sample vibrato support has been improved.  Linear mode
   sample vibratos are now supported perfectly, but modules in Amiga mode
   will not sound exactly correct (but close).
```
"Supported perfectly" is a pretty bold statement considering the waveforms were still wrong at that time. The math used for IT autovibrato also goes back to MikMod 3.0.3 and was fortunately easier to dissect than libxmp's:

```C
                if (s->vibflags & AV_IT) {
                        if ((aout->aswppos>>8)<s->vibdepth) {
                                aout->aswppos += s->vibsweep;
                                vibdpt=aout->aswppos;
                        } else
                                vibdpt=s->vibdepth<<8;
                        vibval=(vibval*vibdpt)>>16;
                        if (aout->mflag) {
                                if (!(mod->flags&UF_LINEAR)) vibval>>=1;
                                aout->main.period-=vibval;
                        }
                } else {
```

1) Each XM autovibrato waveform returns a value on the range [-63,63] (±6 bits).
2) IT autovibrato shifts the depth ([0,63], 6 bits) left 8 bits to implement sweep (correct). (14 bits).
3) These values are multiplied (±20 bits), then shifted right 16 (±4 bits) (and shifted left again in Amiga mode (±3 bits)).
4) Linear periods in MikMod are measured in increments of 2 fine slide steps, so when this is subtracted from the period it translates to a range of about ±32 fine slide steps (when it should be about ±64).

Fixing this seems to be a matter of:

1) Use the correct waveforms for IT.
2) Don't divide the LFO output value for IT (±8 bits). The reduced amplitude seems to have been introduced to compensate for XM autovibrato depth being multiplied in the XM loader, but that's a different issue I don't want to touch at all (since XM autovibrato sounds OK right now).
3) The resulting multiplied value should be ±22 bits, or ±6 bits after right shift, which is the correct linear fine tune steps value in linear mode.
4) This value happens to sound fairly accurate to IT 2.14 when added directly to the period in Amiga mode (pretty much just winging this though), but as mentioned before MikMod linear periods are half of fine slide steps, so it needs to be divided again. This is equivalent to just removing the negation from the `UF_LINEAR` check.
5) IT autovibrato now sounds less incorrect than before.

AliceLR:
@sezero what is your opinion on adding new module flags (`UF_*`)? It looks like there is space for only 5 more of them before the flags field has to be expanded (ABI change), but there are some things that could benefit:

* MOD vibrato bug where it isn't applied on the first tick, resulting in a very messy-sounding vibrato. xmp only applies this to ProTracker/OpenMPT MODs, but MikMod applies it to everything that uses MOD vibrato internally (MED, S3M/old effects IT, several others).
* Some formats update the vibrato LFO position on every tick (IT, GDM, maybe MED) but anything that isn't using the IT vibrato implementation can't do this right now. GDM is using IT vibrato right now but it should have the regular vibrato depth, not IT vibrato depth.
* Some formats update other effects every tick and currently can't right now. MED should do this for volslide and portamento when the MOD-compatible slides flag is off but currently doesn't.

I don't know that new flags are the correct solution for these but I was curious if you had any thoughts regarding that. FWIW barely more than 1/4 of the UNI bytecode space is being used currently and new effects would allow more precise customization of format-specific commands without moving MikMod closer to needing an ABI change. There was also conversation in the libxmp TFMX request issue about how quirk flags are exactly the wrong way to handle some of these format differences... :^)

AliceLR:
I think this patch is about as "ready for review" as it's going to get. There are more minor format-specific changes that could be done but I don't know if they're worth it right now:

* Some players have further slight differences in how they handle certain waveforms.
* Vibrato/panbrello/etc. depth still probably isn't 100% accurate.
* Formats still using `UniPTEffect(0x4,...)` that probably need to be checked at some point: 669, DSMI AMF, DSM, FAR, MTM, STM, STX, ULT. As-is they all should behave exactly the same as before (terrible MOD vibrato).
* GDM uses IT vibrato and should probably not. This patch started while I was working on a separate patch to fix GDM effects... :(

Test modules used to verify that anything changed is at least not more wrong than before: [waveform.zip](https://github.com/sezero/mikmod/files/5566726/waveform.zip)

sezero:
> @sezero what is your opinion on adding new module flags (`UF_*`)? It looks like there is space for only 5 more of them before the flags field has to be expanded (ABI change), but there are some things that could benefit:

An ABI change is not wanted unless it fixes something critical.

However, such changes would be lost welcome if I find the courage and
time to overhaul the library, a 'libmikmod/4' if you will, which would
remove 25 year old legacies (eg. the driver interface, hardware driver
dependencies, windows data types and many other things..)

> (edit: I went ahead and added new UNI effects for the required vibrato changes.)

And, as far as I can see, you introduced no ABI changes, which is good.

sezero (approved):
This should be OK.

I tested very minimally but I trust you and merging it.

Issue #29, opened 2020-11-20
MikMod CLI player is unresponsive, playback lags noticeably behind display.


AliceLR:
I haven't tested the MikMod CLI player very much in Linux but in Windows 10/MSYS2, I've encountered the following bugs consistently:

* When using OSS, the player often takes a second or more to respond to Q or Ctrl+C presses. This isn't an issue with WinMM so it might be a problem with MSYS2 OSS.
* The player display is usually a second or two ahead of playback (both OSS and WinMM).
* When using the WinMM driver, sometimes the end of the track will be cut off (presumably because the player is ahead of the audio playback). This doesn't happen when using the OSS driver.

These bugs can make testing annoying sometimes. I don't know the cause of these but I can look into them further.

AliceLR:
The WinMM audio lag and end of track being cut off issues both disappear when I change the number of buffers in the WinMM driver to 2 instead of 6...

sezero:
I honestly don't know how and can't believe that MSYS builds catch OSS
as an enabled driver. So, I will ignore any MSYS+OSS issues. I can not
test at the moment but OSS on Linux has been behaving very well. If PR
#30 fixes everything else, this can be closed with that.

AliceLR:
> I honestly don't know how and can't believe that MSYS builds catch OSS
> as an enabled driver. So, I will ignore any MSYS+OSS issues.

The header for it is provided by the MSYS2 core packages but I'm not sure which library handles it or how. Since MSYS2 is trying to be most of a Linux environment it makes sense it'd have bindings for OSS.
```
$ pkgfile /usr/include/sys/soundcard.h
msys/msys2-runtime-devel
```

I don't know what exactly configure is deriving `$host_os` from but if it's `uname -o` the result is `Msys`, which isn't detected in configure.ac. The configure script also did not detect WinMM until I manually edited out whatever was disabling it. Doing anything about that seems out of scope for this issue but I think 1) detection of MSYS2 should be fixed (if it's broken) to at the very least allow the WinMM driver, 2) I could go either way re: enabling or disabling OSS; it's not broken outright but it shouldn't be the default driver with this setup.

> I can not test at the moment but OSS on Linux has been behaving very well. If PR #30 fixes everything else, this can be closed with that.

I think that would be OK.

sezero:
> > I can not test at the moment but OSS on Linux has been behaving very well. If PR #30 fixes everything else, this can be closed with that.
> 
> I think that would be OK.

[...]

PR #30, opened 2020-11-20
Add WinMM driver command line options for buffer size and count.

AliceLR:
Adds command line options to configure the number of buffers and size of buffers for the WinMM driver. The default 120ms buffer at 44100Hz works out to 5292 samples, which there definitely did not need to be 6(!) of, so the default buffer count is now 2 (can be configured up to 16 for the off chance someone's dinosaur machine actually needed 6 buffers). The minimum buffer size is 50ms because MikMod's WinMM output gets choppy below that point.

Fixes the WinMM related portions of #29. I don't have any ideas re: the OSS portions except that the MSYS2 OSS driver might just have a lot of latency.

### DIFF
--- a/libmikmod/drivers/drv_win.c
+++ b/libmikmod/drivers/drv_win.c
@@ -58,15 +58,40 @@ typedef DWORD DWORD_PTR;
 #define WAVE_FORMAT_IEEE_FLOAT 0x0003
 #endif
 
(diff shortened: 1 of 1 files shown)


sezero:
Merged this one.  See my notes in #29 and tell me whether it can be closed.

PR #32, opened 2020-11-24
Fix more GDM effects.

AliceLR:
Fixes the following GDM effects:

* Vibrato (`4xy`) and vibrato+volslide (`6xy`): should not have been implemented with IT vibrato. I assigned GDM vibrato a UNI effect opcode and for now it maps to PT vibrato with the tick 0 bug disabled (same as S3M/XM). Specific characteristics of GDM vibrato: no tick 0 update, sine is positive, ramp down actually ramps up (didn't bother supporting—seems like a BWSB bug), all waveforms continue to a new note rather than restarting (didn't bother supporting—seems like a BWSB bug), ramp "down" has a bizarre stair-step nature I didn't bother reproducing (EDIT: this only happens with the SB drivers, not with GUS, so likely a bug/limitation), random waveform is a repeat of square (didn't bother supporting).
* Tremolo (`7xy`): should not have been implemented with S3M tremolo. I assigned GDM tremolo a UNI effect opcode and for now it maps to PT tremolo with the tick 0 bug disabled. Characteristics: same as vibrato. (EDIT: doesn't work with GUS for some reason, tested with SB16 instead)
* Fine vibrato (`14xy`): should not have been implemented with IT fine vibrato. I assigned GDM fine vibrato a UNI effect opcode and for now it maps to S3M fine vibrato. Characteristics: same as vibrato.
* Glissando control (`E3x`), vibrato waveform (`E4x`), set finetune (`E5x`), loop (`E6x`), and tremolo waveform (`E7x`) should all map directly to their Protracker equivalents (`S00` is not supported by BWSB so there's no reason to use the S3M/IT equivalents). Prior to this most of these were being mapped to S3M/IT `Sxy` sub-effect codes(!) which *never* worked and makes me question how well this loader was originally tested. Loop was the only one of these that worked correctly.
* Set global volume (`13xx`) was doubling the parameter value for no good reason.

Other changes:

* Because GDM vibrato/tremolo are more copypaste of the PT versions with different flags and S3M/XM vibrato already do exactly what GDM vibrato needed to do, I combined these into a single function (one for vibrato and one for tremolo). I think it's still a good thing to have opcodes assigned for these since it allows them to be tweaked independently later if needed.
* Protracker tremolo should have the tick 0 bug (just like vibrato). This flag is implemented now but isn't actually used yet because, like vibrato, there are numerous other formats using PT tremolo directly that could be effected.
* I missed this in the big vibrato PR, but S3M fine vibrato was implemented incorrectly and exhibited the PT tick 0 bug when it never should have. It was also halving the depth when it should not have. Tested with Scream Tracker 3.

Test modules:

* [WAVEFORM.GDM.zip](https://github.com/sezero/mikmod/files/5587017/WAVEFORM.GDM.zip): vibrato (`4xy`), fine vibrato (`14xy`), tremolo (`7xy`), vibrato waveform (`E4x`), tremolo waveform(`E7x`).
* [waveform.s3m.zip](https://github.com/sezero/mikmod/files/5587083/waveform.s3m.zip): S3M waveform test including fine vibrato.
* [GLISS.GDM.zip](https://github.com/sezero/mikmod/files/5587040/GLISS.GDM.zip): glissando control (`E3x`). Note that glissando control is not implemented yet in MikMod and both glissando sequences will sound identical, but previously these were getting mapped to opcode 1 and would audibly glitch.
* [FINETUNE.GDM.zip](https://github.com/sezero/mikmod/files/5587033/FINETUNE.GDM.zip): set finetune (`E5x`).
* [GVOLUME.GDM.zip](https://github.com/sezero/mikmod/files/5587063/GVOLUME.GDM.zip): set global volume (`13xx`). Should gradually decrease from 100% to 0% and back. Prior, this would cap at 100% for too long at the start and end due to a doubled parameter.


sezero:
Thanks.

PR #34, opened 2021-05-05
Checks for extra "no loop" marker

neumatho:
In the module "Lost in Germary" by GMaxx (can be found on Modland) have another marker for the samples to indicate that they do not loop. Normal marker is 0x000fffff, but in this module, the marker is 0x00f0ffff. Therefore I have added this extra check.

### DIFF
--- a/libmikmod/loaders/load_669.c
+++ b/libmikmod/loaders/load_669.c
@@ -320,7 +320,7 @@ static BOOL S69_Load(BOOL curious)
 		sample.length=_mm_read_I_SLONG(modreader);
 		sample.loopbeg=_mm_read_I_SLONG(modreader);
 		sample.loopend=_mm_read_I_SLONG(modreader);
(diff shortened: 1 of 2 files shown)


neumatho:
Attached the module.

[Lost In Germany.zip](https://github.com/sezero/mikmod/files/6428213/Lost.In.Germany.zip)

sezero:
Can we change the check to `if (sample.loopend>=0xfffff) sample.loopend=0;`
which is what libxmp does?

neumatho:
Just tested it with your suggesion and that's working fine as well.

AliceLR:
I'm not particularly familiar with MOD player behavior with loopend values >0xFFFFF but this behavior seems like a reasonable default to me.

neumatho:
Should this fix be included or are you still consider it?

sezero:
> Should this fix be included or are you still consider it?

I will apply soon, along with some other patches.

PR #35, opened 2021-05-08
Fixed a little bug in S3M effect A (set speed). 0x80 is also a negati…

neumatho:
Fixed a little bug in S3M effect A (set speed). 0x80 is also a negative number, that's why the equal sign need to be there too. Found a module that sets the speed to 0x80, which made over a 2 minutes silence.

First I though that the bug need to be fixed in the DSMI/AMF loader, but found out that the issue is in fact in the player. Found module is attached. The speed is set at the end of the module.

[Metromania Partie 1.zip](https://github.com/sezero/mikmod/files/6445324/Metromania.Partie.1.zip)

### DIFF
--- a/libmikmod/playercode/mplayer.c
+++ b/libmikmod/playercode/mplayer.c
@@ -1157,7 +1157,7 @@ static int DoS3MEffectA(UWORD tick, UWORD flags, MP_CONTROL *a, MODULE *mod, SWO
 	if (tick || mod->patdly2)
 		return 0;
 
-	if (speed > 128)
+	if (speed >= 128)
 		speed -= 128;
 	if (speed) {
 		mod->sngspd = speed;


sezero:
@AliceLR: Is this good?

AliceLR:
At a glance this patch seems correct to me. Since it affects formats other than DSMI AMF (S3M, IT, GDM, IMF; OctaMED references this effect but bounds it outside of the affected range), I'll do a little bit of testing soon.

AliceLR:
I haven't checked the other formats yet but the behavior of `0x80` being treated as speed 128 appears to be accurate to Dual Module Player 4.00, the MOD player that [DSMI and the AMF format were originally created for](https://github.com/ochrons/dsmi). The notes in the final pattern also seem consistent with it being intended as speed 128 (subjective). OpenMPT also treats this as speed 128.

This doesn't mean this patch is wrong, but if the ST3 and DMP behaviors are different this effect might need to be split.

neumatho:
Does DMP also have silence for over 2 minutes at the end? I just tried with OpenMPT to see how it handles it, and it also sets the tempo to 128, so it seems to be the correct way.

Now we only need to check how the Streamtracker 3 editor handles it, to see if we need to split the command or not. Can you check that?

libmikmod/playercode/mplayer.c
```diff
@@ -1157,7 +1157,7 @@ static int DoS3MEffectA(UWORD tick, UWORD flags, MP_CONTROL *a, MODULE *mod, SWO
 	if (tick || mod->patdly2)
 		return 0;
 
-	if (speed > 128)
+	if (speed >= 128)
 		speed -= 128;
```
AliceLR, line 1161: Finally got around to testing all of the formats I mentioned above. Everything that uses the S3M speed effect expects values 0x80-0xFF to map to speeds 128-255.

S3MEFFECTA compatibility:

| Loader | Player | Effect | Values >=0x80? |
|-|-|-|-|
| load_amf.c	| DMP			| 0x01		| values over 0x80 as speed >=128 (MikMod is wrong)
| load_gdm.c	| BWSB			| 0x0F		| values over 0x80 as speed >=128 (both xmp and MikMod are wrong)
| load_imf.c	| IMAGO Orpheus	| 0x01 (1)		| values over 0x80 as speed >=128 (MikMod is wrong, xmp support broken?)
| load_it.c		| Impulse Tracker	| 0x01 (A)		| values over 0x80 as speed >=128 (MikMod is wrong)
| load_med.c	| OctaMED		| 0x09 (09)	| bounded to 0x20
| load_s3m.c	| Scream Tracker	| 0x01 (A)		| values over 0x80 as speed >=128 (MikMod is wrong)

So, IMO, the correct fix is to just remove both of these lines altogether. Some testing on my end suggests doing this will reproduce the behavior I saw in the trackers/players I tested.

Test files used for GDM, IMF, IT, S3M: [S3MEFFECTA.zip](https://github.com/sezero/mikmod/files/6569379/S3MEFFECTA.zip).

AliceLR (changes requested):

sezero:
Send pa patch (a PR) that overrides this one.

AliceLR:
Going to go ahead and close this, since #42 has been merged.

Issue #9, opened 2020-10-14
STM loader handles empty patterns and the end of song order byte poorly.


AliceLR:
From the STM zero volume samples issue: https://github.com/sezero/mikmod/issues/6#issuecomment-708124148. Essentially, the bounding on the STM loader's loop for reading the order list isn't quite right, and the STM loader also does not correctly handle empty patterns with a valid index between 0-63. Several .STMs rely on this empty pattern behavior.

The MikMod 3.1.9 handling is:
```C
        /* 99 terminates the patorder list */
        while((mh->patorder[t]!=99)&&(mh->patorder[t]<mh->numpat)) {
                of.positions[t]=mh->patorder[t];
                t++;
        }
        if(mh->patorder[t]!=99) t++;
        of.numpos=t;
        of.numtrk=of.numpat*of.numchn;
        of.numins=of.numsmp=31;
```
The MikMod 3.1.10 handling (which seems to be identical to the current handling aside from a missing bounds check) is:
```C
        /* 99 terminates the patorder list */
        while((mh->patorder[t]<=99)&&(mh->patorder[t]<mh->numpat)) {
                of.positions[t]=mh->patorder[t];
                t++;
        }
        if(mh->patorder[t]<=99) t++;
        of.numpos=t;
        of.numtrk=of.numpat*of.numchn;
        of.numins=of.numsmp=31;
```
These are both wrong. The first version allows bytes >99, which is probably the reason for the revision. The second version incorrectly includes 99 in the bound, meaning every track gets a junk order appended after the loop.

The line after the loop that increments `t` in both versions is also suspect. I think this might have been added in an attempt to work around the second bug I mentioned, but the resulting extra order is never valid (it's either 99 or >=`mh->numpat`).

I think the correct handling is something along the lines of:

```C
static BOOL STM_Load(BOOL curious)
{
        BOOL blankpattern=0;
        int t;
...
        /* 99 terminates the patorder list */
        while(mh->patorder[t]<99) {
                /* Patterns >=64 have undefined behavior... */
                if(mh->patorder[t] >= 64) {
                        _mm_errno = MMERR_NOT_A_MODULE;
                        return 0;
                }

                of.positions[t]=mh->patorder[t];

                /* ScreamTracker 2 treats patterns >= the pattern count like blank patterns. */
                if(of.positions[t]>=mh->numpat) {
                        of.positions[t]=mh->numpat;
                        blankpattern=1;
                }

                if(++t == 0x80) {
[...]

PR #38, opened 2021-05-18
Some small fixes for ScreamTracker 2

neumatho:
1. One extra position was always added, which made the player to play the first pattern at the end.
2. Fixed sample loop and detection for invalid values, e.g. N.S. Beat.stm and Batman.stm.

[stm_mods.zip](https://github.com/sezero/mikmod/files/6502044/stm_mods.zip)

### DIFF
--- a/libmikmod/loaders/load_stm.c
+++ b/libmikmod/loaders/load_stm.c
@@ -325,7 +325,6 @@ static BOOL STM_Load(BOOL curious)
 			return 0;
 		}
 	}
-	if(mh->patorder[t]<=99) t++;
 	of.numpos=t;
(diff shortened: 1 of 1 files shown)


sezero:
@AliceLR: can you please review?

AliceLR:
My initial impressions:

* The first fix is OK. It partially fixes #9, which is good. I can submit the rest as a separate patch.
* The second fix looks OK at a glance. I checked libxmp and it has the same behavior as MikMod without this patch, so I'll confirm and then go ahead and merge this (and submit a patch to libxmp if needed).

AliceLR:
I'm going to go ahead and merge this.

PR #39, opened 2021-05-23
Reset channels

neumatho:
If you have set so the module loops instead of stop, then all channels are now cleared. This can e.g. be heard in "7 seconds.s3m", where some chords are hanging.

[7 Seconds.zip](https://github.com/sezero/mikmod/files/6527744/7.Seconds.zip)

### DIFF
--- a/libmikmod/loaders/load_669.c
+++ b/libmikmod/loaders/load_669.c
@@ -320,7 +320,7 @@ static BOOL S69_Load(BOOL curious)
 		sample.length=_mm_read_I_SLONG(modreader);
 		sample.loopbeg=_mm_read_I_SLONG(modreader);
 		sample.loopend=_mm_read_I_SLONG(modreader);
-		if (sample.loopend==0xfffff) sample.loopend=0;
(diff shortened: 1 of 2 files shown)


sezero:
@AliceLR: Can you please review?

sagamusix:
Please please please don't do this. ModPlug Tracker did this a long time ago (cutting channels on loop) but it was the *only* tracker doing that, so modules made in other trackers didn't loop correctly, and modules made in ModPlug didn't play correctly in other trackers. Scream Tracker most definitely doesn't cut channels on loop, and as this module was made with Scream Tracker 3.01, the module would no longer play the same way as it does in ST3 with this patch. Instead, consider the module to be broken. For all we know, this particular module here is unfinished or missing a pattern jump at the end. Just stopping all channels on loop still doesn't make the module sound right, it would sound more correct if in the last pattern there was a jump to pattern 6 or so. But just cutting the playing channels on loop is *wrong*.

neumatho:
Good points there. I will revert these changes on my ported version of MikMod. You may close this pull request and ignore it.

sezero:
Closed.

PR #41, opened 2021-05-31
Add .gitignore.

AliceLR:
This repository currently has no `.gitignore` file, so the output of `git status` ends up being polluted by a large number of generated build system and object files. I've had this sitting around for a while but never got around to submitting it. This is *not* a complete list of generated files, just the ones I've encountered so far in my MSYS2 MikMod development environment, so feedback is welcome. (I'm also not very familiar with the autotools generated files.)


sezero:
Thanks (and welcome back)

PR #51, opened 2021-06-03
Fix DSMI AMF note 0x7f, add support for older AMF 1.0 modules.

AliceLR:
This patch fixes a couple of AMF issues that weren't addressed by #36. These were found here: https://github.com/libxmp/libxmp/issues/350.

* Adds support for older AMF 1.0 modules. This was broken at some point in the past because someone (correctly) realized that not all AMF 1.0 modules have truncated loop start/end values. I had to track down DMP 2.32 to make one of these myself, since the modules libxmp cites have vanished. [reborning.amf.zip](https://github.com/sezero/mikmod/files/6590085/reborning.amf.zip)
* Note 0x7f doesn't duplicate the previous event (like AMF.TXT claims), but instead tells DSMI to retrigger the current playing note in some obscure situations. The best thing to do is to ignore these and (someday) to add support for the Protracker 2.3 instrument change quirk this was trying to emulate.
* Fixes a minor warning introduced by #36.

### DIFF
--- a/libmikmod/loaders/load_amf.c
+++ b/libmikmod/loaders/load_amf.c
@@ -114,7 +114,48 @@ static void AMF_Cleanup(void)
 	track=NULL;
 }
 
-static BOOL AMF_UnpackTrack(MREADER* r)
+/* Some older version 1.0 AMFs contain an anomaly where the sample length is
+ * a DWORD, the loop start is a WORD, and the loop end is missing. Later AMF
+ * 1.0 modules and up contain all three as DWORDs, and earlier versions have
+ * all three as WORDs. This function tries to detect this edge case in the
+ * instruments table. This should only be called on 1.0 AMFs.
+ */
+static BOOL AMF_ScanV10Instruments(MREADER *r, unsigned int numins)
+{
+	SLONG resetpos;
+	BOOL res = 0;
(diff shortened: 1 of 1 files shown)


libmikmod/loaders/load_amf.c
```diff
@@ -114,7 +114,48 @@ static void AMF_Cleanup(void)
...
-static BOOL AMF_UnpackTrack(MREADER* r)
+/* Some older version 1.0 AMFs contain an anomaly where the sample length is
+ * a DWORD, the loop start is a WORD, and the loop end is missing. Later AMF
+ * 1.0 modules and up contain all three as DWORDs, and earlier versions have
+ * all three as WORDs. This function tries to detect this edge case in the
+ * instruments table. This should only be called on 1.0 AMFs.
+ */
+static BOOL AMF_ScanV10Instruments(MREADER *r, unsigned int numins)
+{
+	LONG resetpos;
```
sezero, line 125: This should be `SLONG`

AliceLR: Whoops, that should be fixed now. Since MikMod uses Windows.h types everywhere I assumed that was one of them, but I was wrong!

PR #55, opened 2021-09-23
STM patterns >=numpat should be handled as blank patterns.

### LINKED ISSUE #9: STM loader handles empty patterns and the end of song order byte poorly.

AliceLR:
In Screamtracker 2, patterns >=numpat and <64 are always valid and treated as blank patterns. MikMod was preserving the out of bounds pattern values but not allocating a blank pattern for them, resulting in these patterns being skipped. Several STM modules use these blank patterns at the end of their order list, presumably to act as a break so the user can switch modules, or to let samples finish.

Fixes #9.

### DIFF
--- a/libmikmod/loaders/load_stm.c
+++ b/libmikmod/loaders/load_stm.c
@@ -225,7 +225,7 @@ static UBYTE *STM_ConvertTrack(STMNOTE *n)
 	return UniDup();
 }
 
-static BOOL STM_LoadPatterns(void)
+static BOOL STM_LoadPatterns(int pattoload)
 {
 	int t,s,tracks=0;
 
@@ -233,7 +233,7 @@ static BOOL STM_LoadPatterns(void)
 	if(!AllocTracks()) return 0;
 
 	/* Allocate temporary buffer for loading and converting the patterns */
-	for(t=0;t<of.numpat;t++) {
+	for(t=0;t<pattoload;t++) {
 		for(s=0;s<(64U*of.numchn);s++) {
 			stmbuf[s].note   = _mm_read_UBYTE(modreader);
 			stmbuf[s].insvol = _mm_read_UBYTE(modreader);
@@ -254,6 +254,8 @@ static BOOL STM_LoadPatterns(void)
 
 static BOOL STM_Load(BOOL curious)
 {
+	BOOL blankpattern=0;
+	int pattoload;
 	int t;
 	ULONG samplestart;
 	ULONG sampleend;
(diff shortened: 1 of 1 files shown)


sezero:
Thanks.

PR #57, opened 2021-10-03
Mod plugin packed

neumatho:
Added support for depacking ADPCM4 samples used in MOD Plugin.

Note that MOD Plugin also can pack samples in S3M, IT and MOD files, but I do not have any of such modules, so only support in XM files has been added.

You can get a lot of modules at modland here: [https://modland.com/incoming/workshop/Modpluginpacked/]


sezero:
@AliceLR: is this good?

sezero:
Rebased properly.

PR #58, opened 2021-10-03
Fix ASYLUM Music Format restart byte.

AliceLR:
See: https://github.com/libxmp/libxmp/pull/486. This fixes misc. fields in the ASYLUM Music Format header that MikMod previously was ignoring.


sezero:
Thanks.

PR #59, opened 2021-10-03
Detect more MOD variants: TDZ#, FA04/6/8, LARD, NSMS.

AliceLR:
See: https://github.com/libxmp/libxmp/pull/484.

This adds support for more MOD magics that MikMod didn't support before. This should just about cover all of the trivial MOD magics I know of that MikMod doesn't support already.
* `TDZ#` where `1 <= # <= 3` is TakeTracker's magic for modules with between 1 and 3 channels. These are fairly rare but their existence is easy to verify with TakeTracker.
* `FA04`/`FA06`/`FA08` are MOD magics from the Atari tracker Digital Tracker. They only differ from regular MODs in that they have an extra 4 bytes after the magic of which the purpose isn't obvious. I don't have an Atari Falcon or an emulator setup to test this, so I just made MikMod skip them. They have been commented indicating this. (libxmp and OpenMPT need to be patched likewise.)
* `LARD` and `NSMS` are weird one-off MOD magics for `judgement_day_gvine.mod` and `kingdomofpleasure.mod`, respectively.

This is very low priority, so marking this as a draft until someone figures out what the deal is wrt the Digital Tracker extra bytes.

### DIFF
--- a/libmikmod/loaders/load_mod.c
+++ b/libmikmod/loaders/load_mod.c
@@ -87,8 +87,10 @@ static CHAR fasttracker[] = "Fasttracker";
 static CHAR octalyser[] = "Octalyser";
 static CHAR oktalyzer[] = "Oktalyzer";
 static CHAR taketracker[] = "TakeTracker";
+static CHAR digitaltracker[] = "Digital Tracker MOD";
 static CHAR orpheus[] = "Imago Orpheus (MOD format)";
 static CHAR modsgrave[] = "Mod's Grave";
+static CHAR unknown[] = "Unknown tracker MOD";
 
 static MODULEHEADER *mh = NULL;
 static MODNOTE *patbuf = NULL;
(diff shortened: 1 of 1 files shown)


AliceLR:
After checking some Digital Tracker versions in an Atari Falcon emulator I managed to come up with the following (IMO fairly reasonable) observations. tl;dr the extra Digital Tracker bytes are intentional and they are also most likely safe to disregard.

Why they're intentional:

* `FA06` and `FA08` MODs can easily be saved in Digital Tracker <=2.01. Later versions switch to DTM.
* Regardless of pattern data, the bytes `00h 40h 00h 00h` are always present after the magic.
* There's a minor anomaly: the ModLand `FA08` modules are the exact length you'd expect if they were `8CHN` modules, suggesting there shouldn't be 4 extra bytes anywhere in the module. However, all `FA08` modules found in Digital Tracker releases and all `FA06`/`FA08` modules saved are consistently the size you'd expect from a `6CHN` or `8CHN` with 4 extra bytes. I think the extra length of the ModLand modules might have been noticed by someone or a utility and truncated off.

Why they can be ignored (and probably what they were meant for):

* Documentation for early versions of Digital Tracker advertises that you can change the "base" of the module in their variant, unlike regular MOD. The editor UI makes it obvious that this means you can set the samples (globally) to be either 8-bit or 16-bit, and one of 8363Hz, 12500Hz, and 25000Hz. These UI widgets, and the UI widget that eventually allows you to change the pattern row count, are unusable prior to the switch to DTM.
* The UI widgets and the default values of `00h 40h 00h 00h` together suggest that the four bytes were intended to be something along the lines of: pattern row count (big endian word=64), sample bits (0=8-bit, 1=16-bit), sample rate (0=8363Hz, 1=12.5kHz, 2=25kHz).
* If the values are edited in a hex editor and imported into a >=2.02 version of Digital Tracker, they don't change any of these features: the row count of every pattern will still be 64, the samples will still be 8-bit, and the sample rate will be 8400.
* Since they can't be changed in versions that save `FA0x` modules and are ignored in versions after the switch to `DTM`, they are effectively unusable and can *probably* be ignored altogether.
* I haven't been able to load any modules (including the hex edited modules) in DT <=2.01 due to crashes in the emulator. I haven't been able to verify these fields also do nothing in these versions either because of this, but I feel like there's probably a fair chance of that.

Going to go ahead and mark this ready for review.

sezero:
Thanks.

PR #63, opened 2021-11-04
Envelope fix

neumatho:
Two fixes in this pull request:

First fix will prevent last used sampled in a channel to be played, if afterwards an illegal instrument is set. Now nothing is played.

The second fix is a big one. Almost totally rewrote the envelope handling, so it behave like FastTracker II does, specially when used together with XM effect L. Inspiration to this fix is from a translation of the original FastTracker II code to C, which can be found here: [Replayer source to FT II Clone](https://github.com/8bitbubsy/ft2-clone/blob/master/src/ft2_replayer.c). It is special the updateChannel() function I looked at.

These two fixes make Ebony Owl Netsuke.xm to play correctly.

[Ebony Owl Netsuke.zip](https://github.com/sezero/mikmod/files/7476168/Ebony.Owl.Netsuke.zip)

### DIFF
--- a/libmikmod/playercode/mplayer.c
+++ b/libmikmod/playercode/mplayer.c
@@ -349,43 +349,78 @@ static SWORD DoPan(SWORD envpan,SWORD pan)
 	return (newpan<PAN_LEFT)?PAN_LEFT:(newpan>PAN_RIGHT?PAN_RIGHT:newpan);
 }
 
-static SWORD StartEnvelope(ENVPR *t,UBYTE flg,UBYTE pts,UBYTE susbeg,UBYTE susend,UBYTE beg,UBYTE end,ENVPT *p,UBYTE keyoff)
+static void StartEnvelope(ENVPR *t,UBYTE flg,UBYTE pts,UBYTE susbeg,UBYTE susend,UBYTE beg,UBYTE end,ENVPT *p,UBYTE keyoff,SWORD defaultvalue)
 {
 	t->flg=flg;
 	t->pts=pts;
 	t->susbeg=susbeg;
 	t->susend=susend;
+	t->susactive=0;
 	t->beg=beg;
 	t->end=end;
 	t->env=p;
 	t->p=0;
-	t->a=0;
-	t->b=((t->flg&EF_SUSTAIN)&&(!(keyoff&KEY_OFF)))?0:1;
+	t->index=1;
 
 	if (!t->pts) { /* FIXME: bad/crafted file. better/more general solution? */
-		t->b=0;
-		return t->env[0].val;
+		t->index=0;
+		t->lastvalue=defaultvalue;
+	} else {
+		/* Imago Orpheus sometimes stores an extra initial point in the envelope */
+		if ((t->pts>=2)&&(t->env[0].pos==t->env[1].pos))
+			t->index++;
+
+		/* Fit in the envelope, still */
+		if (t->index >= t->pts)
+			t->index = t->pts;
+
+		t->lastvalue = t->env[t->index - 1].val;
 	}
 
 	/* Ignore junk loops */
-	if (beg > pts || beg > end)
+	if (loopbeg > pts || loopbeg > loopend)
 		t->flg &= ~EF_LOOP;
 	if (susbeg > pts || susbeg > susend)
 		t->flg &= ~EF_SUSTAIN;
+}
 
-	/* Imago Orpheus sometimes stores an extra initial point in the envelope */
-	if ((t->pts>=2)&&(t->env[0].pos==t->env[1].pos)) {
-		t->a++;
-		t->b++;
+/* Set envelope tick to the position given */
+static void SetEnvelopePosition(ENVPR *t, ENVPT *p, SWORD pos)
+{
+	if (t->pts > 0)	{
+		BOOL found = 0;
+		UWORD i;
+
+		for (i = 0; i < t->pts - 1; i++) {
+
+			if ((pos >= p[i].pos) && (pos < p[i + 1].pos)) {
+				t->index = i + 1;
+				t->p = pos;
+
+				found = 1;
+				break;
+			}
+		}
+
+		/* If position is after the last envelope point, just set
+		   it to the last one */
+		if (!found) {
+			t->index = t->pts;
+			t->p = p[t->index - 1].pos;
+		}
+
+		t->lastvalue = InterpolateEnv(t->p, &t->env[t->index - 1], &t->env[t->index]);
 	}
+}
 
-	/* Fit in the envelope, still */
-	if (t->a >= t->pts)
-		t->a = t->pts - 1;
-	if (t->b >= t->pts)
-		t->b = t->pts-1;
-
-	return t->env[t->a].val;
+/* Set the panning envelope tick to the position given */
+static void SetPanningEnvelopePosition(MODULE *mod, INSTRUMENT *i, ENVPR *t, ENVPT *p, SWORD pos)
+{
+	/* Because of a bug in FastTracker II, only the panning envelope
+	   position is set if the volume sustain flag is set. Other players
+	   may set the panning all the time */
+	if (!(mod->flags & UF_FT2QUIRKS) || (i->volflg & EF_SUSTAIN))
+		SetEnvelopePosition(t, p, pos);
 }
 
 /* This procedure processes all envelope types, include volume, pitch, and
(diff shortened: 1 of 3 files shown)


sezero:
@AliceLR please have a look when you can.

AliceLR:
> The second fix is a big one. Almost totally rewrote the envelope handling

Changes that affect multiple formats like this with no regression testing are, as sezero can confirm, a *significant* contributor to why MikMod's format accuracy has been in such a sad state for a long time. How many modules have you tested this patch against? How many are XM? IT? IMAGO Orpheus?

> First fix will prevent last used sampled in a channel to be played, if afterwards an illegal instrument is set. Now nothing is played.

I think this is another thing that can vary significantly between formats (maybe @sagamusix can confirm...).

sagamusix:
There are significant differences between IT and XM style envelope handling, so just taking an XM player's code here is almost guaranteed to break other formats unless they were already broken to begin with.

> I think this is another thing that can vary significantly between formats (maybe @sagamusix can confirm...).

I don't have tests handy but IIRC IT in instrument mode will disagree with that due to how the instrument to sample mapping works.

neumatho:
I tried to be very careful when doing this change, because I know it may broke something. Did a lot of debugging on multiple modules. I have tested with some modules, but not that many. Will test more. Ofcourse it would be good, if there is some real good testing modules in all formats where you can easily hear how it should sound and if it is broken.

I did not just copy the XM envelope function, but tried to apply what this did in some situation with the original MikMod code. The only difference I could see out of MikMod code between XM and IT, is that XM has one sustain point, while IT has two (begin/end). This part is kept in my changes. Are there other differences I need to be aware of?

And while writing this, there is one thing I need to check. I'm not sure if there is made interpolation when running between multiple sustain points. Will check that and commit any changes if needed.

neumatho:
It seems there are some issues with IT. The strange thing is that my C# port and the original MikMod code do not sound similar, which is very strange. I'm looking at it.

neumatho:
After many hours of analyse and debugging, I finally found why there are differences. It has nothing to do with this patch. In the module "A Life In Termoil.it", there is played a sequence in 4 voices at the beginning. In the original MikMod code, only the first 2 can be heard, even when the code parses them, while my port of MikMod all 4 can be heard. I then found out, that the last two voices are marked as surround channels (via S9x effect). If I disable surround (via DMODE flag), then all 4 channels are played in MikMod. So there seem to be an issue with surround mixing in MikMod.

Back to my patch. This patch does not sound correctly on IT modules, so I think I will make two separate envelope handlers for XM and IT. I then need to figure out exactly how it works. Do any of you know what IMAGO Orpheus uses? XM, IT or its own?

sagamusix:
In general no two trackers from that period are alike because nobody really wrote done in a fully detailed way how their engine works. However, Imago behaves more like FT2 than Impulse Tracker when it comes to envelope handling, so as long as mikmod doesn't have any specialized handling for envelopes in IMF yet, it's probably best to follow what XM does.

neumatho:
I have now separated the envelope handlers into two different methods. The main difference between the XM and IT way of handling envelopes, is that XM process the first point before it checks for sustain/looping, while IT do it for each tick. Lets take an example. We have a simple envelope of two points:

```
Pos  Value
 0    256
12      0
```

The loop has both a start and end on point 0. For XM, the volume will start at 256 and then fade down to 0 in 12 ticks and move to the next point. Because the new point is after the loop, the loop is ignored, so it keep the volume at 0.

For IT, it will keep the volume at 256 the whole time, because it keep looping at the first point.

For the illegal instrument fix, MikMod has always made a test for out-of-bounds instruments, it just forgot to clear some variables, which made it play the last used sample on that sample, whatever that was. So the fix, is just to clear those variables.

sezero:
Rebased this to current master. Reviews? Comments?

sezero:
@AliceLR: What is your final take in this? Reviews, comments?

neumatho:
Just wanna say, that libxmp has the same problems.

libmikmod/playercode/mplayer.c
```diff
@@ -3015,13 +3120,18 @@ static void pt_Notes(MODULE *mod)
 				break;
 			case UNI_INSTRUMENT:
 				inst=UniGetByte();
-				if (inst>=mod->numins) break; /* safety valve */
+				if (inst>=mod->numins) {
+					a->main.i=NULL;
+					a->main.s=NULL;
+					a->main.sample=-1;
+					break; /* safety valve */
+				}
```
AliceLR, line 3128: I don't have any useful examples right now but I *think* this might be completely format-dependent. Judging just on how libxmp does this (not necessarily correct, but generally pretty good):

* MOD: completely cuts the playing note.
* XM/other XM-likes: sets the volume to 0 but continues playing the current note.
* S3M/other S3M-likes: ignores the invalid instrument, continues playing the current note.
* IT (sample mode): sets the volume to 0 but continues playing the current note.
* IT (instrument mode): ignores the invalid instrument, continues playing the current note.
* MED: completely cuts the playing note.

If the above behavior descriptions are accurate, this change *could* be at the expense of S3M, XM, and IT accuracy.

libmikmod/playercode/mplayer.c
```diff
@@ -2806,30 +2909,32 @@ static void pt_UpdateVoices(MODULE *mod, int max_volume)
...
 		if (i && ((aout->main.kick==KICK_NOTE)||(aout->main.kick==KICK_ENV))) {
-			if (aout->main.volflg & EF_ON)
-				envvol = StartEnvelope(&aout->venv,aout->main.volflg,
-				  i->volpts,i->volsusbeg,i->volsusend,
-				  i->volbeg,i->volend,i->volenv,aout->main.keyoff);
-			if (aout->main.panflg & EF_ON)
-				envpan = StartEnvelope(&aout->penv,aout->main.panflg,
-				  i->panpts,i->pansusbeg,i->pansusend,
-				  i->panbeg,i->panend,i->panenv,aout->main.keyoff);
+			if (aout->main.volflg & EF_ON) {
```
AliceLR, line 2912: The changes in this section look 100% correct, I would merely recommend staying close to the old text wrapping where possible (I really hope no one relies on 96-char wrapping though...).

libmikmod/playercode/mplayer.c
```diff
@@ -1613,21 +1719,18 @@ static int DoXMEffectL(UWORD tick, UWORD flags, MP_CONTROL *a, MODULE *mod, SWOR
 	dat=UniGetByte();
 	if ((!tick)&&(a->main.i)) {
 		INSTRUMENT *i=a->main.i;
-		MP_VOICE *aout;
+		MP_VOICE *aout = &mod->voice[channel];
```
AliceLR, line 1722: Not sure what exactly is going on with MikMod's handling of channel assignment; is there a reason this was changed from `a->slave`?

AliceLR (commented):
So far I've reviewed everything but `ProcessEnvelope` and cohorts, and I don't think there are many issues of note. And yes, libxmp's envelope code leaves a bit to be desired, there's already an issue open for *part* of this I think :)

libmikmod/playercode/mplayer.c
AliceLR: Never mind, the text wrapping in other parts of this file are already a mess.

libmikmod/playercode/mplayer.c
neumatho: It is a while ago, since I made this patch, so I cannot exactly remember. I think it is because a->slave is null, since it is a XM effect, so I need to get the voice in a different way.

sezero:
Rebased to mainstream with build fixes (no TRUE/FALSE..)

AliceLR:
Will rebase/patch this for #75 once it's in.

sezero:
> Will rebase/patch this for #75 once it's in.

PING?

sezero:
I rebased this to current git HEAD: compile-tested _only_. Commit 0255e920c (#75)
was in the way, so I am _not 100% sure_ I did it correctly: @AliceLR, @neumatho:
please verify and re-review for the fate of this P/R.

AliceLR:
I don't have much useful to add here right now because this is a large patch and I'm not deeply familiar with how the original players handle envelopes.

Sorry to push this off again, but libxmp has envelope work scheduled for 4.7.2, so I'd rather avoid LGPL envelope code for now and port whatever I find with libxmp here later.

PR #66, opened 2021-12-07
Add cmake rule to disable shared library builds for libmikmod

davidgfnet:
The cmakefile has a rule to enable/disable static builds, which is awesome! However, in some platforms, might not be even possible to generate a shared object (or at least not trivially). For this, it is helpful to have a knob to disable said build.

This PR adds a knob for this, which is enabled by default thus not changing the default behaviour.

### DIFF
--- a/libmikmod/CMakeLists.txt
+++ b/libmikmod/CMakeLists.txt
@@ -170,6 +170,10 @@ SET (ENABLE_RPATH 1 CACHE BOOL "Whether to use an rpath when linking" )
 SET (ENABLE_SHARED 1 CACHE BOOL "Whether to build the shared library" )
 SET (ENABLE_STATIC 1 CACHE BOOL "Whether to build the static library" )
 
+IF (NOT ENABLE_STATIC AND NOT ENABLE_SHARED)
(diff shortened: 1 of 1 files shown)


sezero:
Thanks.  Maybe do this too, though (on top of your patch):
```diff
diff -u a/libmikmod/CMakeLists.txt b/libmikmod/CMakeLists.txt
--- a/libmikmod/CMakeLists.txt
+++ b/libmikmod/CMakeLists.txt
@@ -170,6 +170,10 @@
 SET (ENABLE_SHARED 1 CACHE BOOL "Whether to build the shared library" )
 SET (ENABLE_STATIC 1 CACHE BOOL "Whether to build the static library" )
 
+if(NOT ENABLE_STATIC AND NOT ENABLE_SHARED)
+    message(FATAL_ERROR "Both static and shared builds got disabled. You must enable at least one of them.")
+endif()
+
 ### This is to set the RPATH correctly, so when installed
 ### under a prefix the executables will find the libraries.
 ### See:  http://www.cmake.org/Wiki/CMake_RPATH_handling
```

davidgfnet:
Absolutely! Didnt think about this indeed! Thanks for considering the PR, really appreciated!

sezero:
Thanks, applied.

PR #79, opened 2025-01-18
Updating the web version with the latest modifications and updating missing TypeScript definitions

carlosrafaelgn:


sezero:
This is in. Thanks!

PR #91, opened 2026-07-13
Add bounding checks for DSMI AMF channel remap table.

AliceLR:
The indices in this table corresponding to real channels should also point to real channel numbers.

Verified that Test6.amf, reborning.amf, and the tribal zone.amf still load after this fix. This feature is only used by format versions 0.9 and 1.0.


sezero:
Verified to work: Thanks. Please mark as ready when you feel it's finished.

AliceLR:
Verified `of.numchn` never changes after assignment and that all of the DSMI AMF modules on Modland still load (probably correctly). I think this is OK.

sezero:
Thanks.

sezero:
How is libxmp affected?

AliceLR:
> How is libxmp affected?

libxmp doesn't implement the remapping table at all, so it's fine.

